PR #16, opened 2012-09-11
api documentation first run

antis81:
started javadoc style documentation frame for GitWrap and some basic functions

### DIFF
--- a/src/Libs/GitWrap/BackgroundClone.h
+++ b/src/Libs/GitWrap/BackgroundClone.h
@@ -24,6 +24,12 @@
 namespace Git
 {
 
+    /**
+     * @brief The BackgroundClone class provides functionality to clone a repository in a background thread.
+     *
+     * @ingroup GitWrap
+     * @{
+     */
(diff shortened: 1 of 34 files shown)


scunz:
I've merged this, but I might be temporarily reverting it, if it causes too much trouble in rebasing `scnuz/topic/GitWrap_CleanUps`.

scunz:
To check this a bit better, I just instructed the Jenkins to run Doxygen on the [`src/Libs` subdir](http://docs.macgitver.org) on every checkin.

antis81:
Thanks. I wasn't aware of Jenkins, but cool when it is there. The doxyfile is missing though.

scunz:
Both Jenkins instances for MGV are not public, neither is the Doxyfile. I will make it public as soon as I got some additional scripts around it.
One Jenkins just builds libgit2+mgv - and if that succeeds, commits a Submodule update to GH.
The other one builds LibGit2, Qt4 and Qt5 from git and builds MGV on top of that with GCC on Qt4.8.2; GCC on Qt5-HEAD and CLANG on Qt 4.8 HEAD

PR #19, opened 2012-09-11
implemented method Repository::discover() (take 2) 

antis81:
fixed the discover method to use fromUtf8 / toUtf8 paths. could you please check it once more, as I am not 100% sure if it is correct. on the other hand, we'll notice it, if the first utf-8 path is used to discover a repo :smile:.

### DIFF
--- a/src/Libs/GitWrap/Repository.cpp
+++ b/src/Libs/GitWrap/Repository.cpp
@@ -36,6 +36,8 @@
 #include "RevisionWalkerPrivate.h"
 #include "Submodule.h"
 
+#include <QFile>
+
 namespace Git
 {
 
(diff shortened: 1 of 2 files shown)


src/Libs/GitWrap/Repository.cpp
```diff
@@ -124,8 +124,21 @@
...
+        return Repository( new RepositoryPrivate( repo ) );
+    }
+
+    QString Repository::discover(const QString& startPath, bool acrossFs, const QStringList& ceilingDirs)
+    {
+        QByteArray repoPath(GIT_PATH_MAX, Qt::Uninitialized);
+        QByteArray joinedCeilingDirs = ceilingDirs.join(QChar::fromLatin1(GIT_PATH_LIST_SEPARATOR)).toUtf8();
+
+        git_repository_discover( repoPath.data(), repoPath.length()
+                                 , QString(startPath).toUtf8().constData()
```
scunz, line 136: Actually there's no need to QString() it (I assume it was either in error or to cast way constness). However, toUtf8() produces a temporary QByteArray which stays alive until after the function call. C++ is nice, at times :-)

scunz:
I'll change that QString() when i come along that code next time.

src/Libs/GitWrap/Repository.cpp
antis81, line 136: Thanks! It actually came because of I forgot to remove :smile:. Feel free to fix :heart:.

PR #22, opened 2012-09-15
WorkingTree: Make auto-update work

scunz:


scunz:
`libgit2` doesn't update its view of the _index_ before diff'ing against the index, so _any_ updates on _any_ view involving the index (That is all of the Working Tree) will have to wait until this is fixed.

PR #24, opened 2012-09-16
More History implementation

scunz:
- Mostly makes the "Diff View" work

### DIFF
--- a/src/Modules/History/HistoryDiff.cpp
+++ b/src/Modules/History/HistoryDiff.cpp
@@ -52,7 +52,27 @@ HistoryDiff::HistoryDiff()
 	setLayout( l );
 }
 
+void HistoryDiff::setRepository( Git::Repository repo )
+{
+	mRepo = repo;
+}
+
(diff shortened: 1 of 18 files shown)


scunz:
Merging this in; next step is to get support for short-cuts and invisible actions from HIC... Didn't I _just recently_ say, that hic requires a rewrite...

src/Modules/History/HistoryDiff.cpp
```diff
@@ -29,6 +49,227 @@
...
+	mDiffToActions.insert( DTT_Tag, mToolbar->addWidget( mDiffToTag ) );
+
+	mDiffToParent = new QComboBox;
+
+	mParentsModel = new ShortCommitModel( this );
+	mDiffToParent->setModel( mParentsModel );
+	mParentsList = new QTreeView;
+	mParentsList->setHeaderHidden( true );
+	mParentsList->setRootIsDecorated( false );
+	mParentsList->header()->setResizeMode( QHeaderView::ResizeToContents );
```
scunz, line 81: Qt5 barks on this line...

PR #29, opened 2012-09-30
Create a Config Generator

scunz:
@antis81,
Could you review this change? Not sure if it is worth the work at all.

What I wanted to do, is to reduce the configuration mess that any decently complex application involves. This is partly inspired by KDE.
### The current situation is

LibConfig maintains a `QSettings` object where anybody can read and write. The `ConfigUser` class provides an observer-pattern style tool that lets the user inherit `ConfigUser` and specify a sub-tree of the `QSettings` object we're interested in. This is nice, as long as config is used only in one place (Which is actually never true: Config-Dialog and Config-Usage are the minimum in theory).
### My goal is

Write down the configuration entries in an XML File. Maybe even amend them with documentation. An XML transformer will read that and create a singleton class out of it. This singleton associates itself with the global configuration from LibConfig and implements a `ConfigUser` of its own. It provides simple static getters and setters for every value defined in XML.

What do you think about this?


scunz:
Can't test the Qt5 changes, locally. Jenkins will test them when this is merged...

scunz:
LOL, this happens if you force push too fast and jenkins can't catch up:

```
[ 54%] Building CXX object src/Libs/Widgets/CMakeFiles/Widgets.dir/FlatTreeComboBox.cpp.o
c++: internal compiler error: Terminated (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://bugs.gentoo.org/> for instructions.
make[2]: *** [src/Libs/Widgets/CMakeFiles/Widgets.dir/FlatTreeComboBox.cpp.o] Error 4
```

scunz:
This seems to be feature complete now. The generated code looks like:

``` c++
// HistoryConfig.hpp:
class HistoryConfig : public QObject, private ConfigUser
{
    Q_OBJECT
public:
    HistoryConfig( QObject* parent = 0 );
    ~HistoryConfig();

private:
    void read();

public: // History/SplitLayout
    static quint32 splitLayout();
    static void setSplitLayout( quint32 value );

public: // History/Details/List
    static QString detailsList();
    static void setDetailsList( QString value );
signals:
    void detailsListChanged( QString newValue );

protected:
    void configChanged( const QString& subPath, const QVariant& value );

private:
    static HistoryConfig* sSelf;

    quint32 mValueSplitLayout;
    QString mValueDetailsList;
};

// HistoryConfig.cpp:

HistoryConfig::HistoryConfig( QObject* parent )
    : QObject( parent )
    , ConfigUser( "History" )
    , mValueSplitLayout( 0 )
{
    Q_ASSERT( sSelf == NULL );
    sSelf = this;
    read();
}

HistoryConfig::~HistoryConfig()
{
    Q_ASSERT( sSelf != NULL );
    sSelf = NULL;
}

HistoryConfig* HistoryConfig::sSelf = NULL;

void HistoryConfig::read()
{
    mValueSplitLayout = configGet< quint32 >( "SplitLayout", 0 );
    mValueDetailsList = configGet< QString >( "Details/List", QString::fromUtf8( "#" ) );
}

quint32 HistoryConfig::splitLayout()
{
    Q_ASSERT( sSelf != NULL );
    return sSelf->mValueSplitLayout;
}

void HistoryConfig::setSplitLayout( quint32 value )
{
    sSelf->configSet( "SplitLayout", value );
}

QString HistoryConfig::detailsList()
{
    Q_ASSERT( sSelf != NULL );
    return sSelf->mValueDetailsList;
}

void HistoryConfig::setDetailsList( QString value )
{
    sSelf->configSet( "Details/List", value );
}

void HistoryConfig::configChanged( const QString& subPath, const QVariant& value )
{
    if( subPath == QLatin1String( "SplitLayout" ) )
    {
        mValueSplitLayout = value.value< uint >();
        return;
    }

    if( subPath == QLatin1String( "Details/List" ) )
    {
        mValueDetailsList = value.value< QString >();
        emit detailsListChanged( mValueDetailsList );
        return;
    }
}
```

That should be fine so far. Still missing support for lots of Data-Types, though.

scunz:
I've just merged this. To me it seems working and it blocks me from getting the sub modules right...

PR #42, opened 2012-10-04
GitWrap as submodule

scunz:


scunz:
retest this please

scunz:
This damned regex doesn't like me, i assume...

PR #45, opened 2012-10-21
Submodule UI Module: The next step

antis81:
Finally I found some time to give the submodule view another try. I kept the "Novice.GUI.xml" file  local, so nothing is visible yet.

Features:
- reading of submodules in bare and normal repositories works
- ToolTip shows submodule details (URL and local relative path)

Missing:
- Because the IconProvider is not finished yet, an empty QIcon is used as a placeholder.
- As well,   submodules are not read recursively.
- The current submodule status.
- instead of the ID something like "git describe" should be shown (meaning: a tag, if there is one)

Questions:
- What was the intention of the (unused) visitor pattern in "readSubmodules()"? I'd suggest to leave it out, until we really need such a thing and do the right thing then.
- To make it work, I really had to take out the "LIBGIT2_IMPROVED" workaround. Otherwise nothing will be shown. Seems very ok to me. What do you think.

### DIFF
--- a/src/Modules/Submodules/CMakeLists.txt
+++ b/src/Modules/Submodules/CMakeLists.txt
@@ -14,6 +14,7 @@ SET( SRC_FILES
     SubmodulesCreateEditDlg.cpp
     SubmodulesModule.cpp
     SubmodulesView.cpp
+    SubmodulesDelegate.cpp
 )
 
 SET( HDR_FILES
@@ -21,6 +22,7 @@ SET( HDR_FILES
     SubmodulesCreateEditDlg.h
     SubmodulesModule.h
     SubmodulesView.h
+    SubmodulesDelegate.h
 )
 
 SET( UI_FILES
--- a/src/Modules/Submodules/SubmodulesDelegate.cpp
+++ b/src/Modules/Submodules/SubmodulesDelegate.cpp
@@ -1,0 +1,100 @@
+#include "SubmodulesDelegate.h"
+
+#include <QAbstractItemView>
+#include <QPainter>
+#include <QToolTip>
+#include <QHelpEvent>
+
+#include "libGitWrap/ObjectId.h"
+#include "libGitWrap/Submodule.h"
+
+
+SubmodulesViewDelegate::SubmodulesViewDelegate( QObject* parent )
+    : QItemDelegate( parent )
+{
+}
+
+void SubmodulesViewDelegate::paint( QPainter* painter, const QStyleOptionViewItem& option,
+                                    const QModelIndex& index ) const
+{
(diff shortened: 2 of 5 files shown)


scunz:
Hi @antis81, I've been very inactive the last few weeks - and it will probably stay like that for the rest of the month.

However, looking at this, I tend to say: ship it; but actually not because I think that it's the right way to do it; but rather: This change will not make the submodule view more broken than it already is.

The reasoning behind is:
- The Submodule-View is very different from anything else that MacGitver will have to implement. We need to supply some more core-services to make it actually work as we want. These need support from the libHeaven and from MGV's very own core. Actually, we need to Context-Switch all other views depending on the SM-View's current selection (or whatever trigger we invent to switch between submodules).
- The FS-Watcher implemented in MGV's core is insufficient. We cannot get really accurate information on submodule changes.
- On top of that, libgit2 refuses to reread the submodules, once it has them loaded. That's why we _need_ to open the repository a second time to get up-to-date submodule information. I'm not sure, whether @arrbee meanwhile changed that, has thought about it or even is aware of it. Might as well be true, that my knowledge has run out of date... Libgit2 had lots of activity this month, which I couldn't follow too deeply either.
- If we really want to support nested submodules, we would have to reinvent the read-logic of the submodule view from scratch. I think, that the core must keep track of all repositories associated with the tree of submodules (And we have to build a tree out of that information). However this is far beyond your changes here.
- As for "unused visitor pattern", well. It actually isn't unused. `readSubmodules()` is supposed to be called on every change to the submodule tree (which would at least include .gitmodules changes and .git/config changes). The submodule view _should_ then remove any submodules that are no longer part of the root-repository...

antis81:
Thanks for reviewing this quickly. Well, I'll merge it then for hopefully more changes to come later.

Hm, I think it is too early to bother about how to get all information about submodule changes. First, like you mentioned, the view should become rather a central point managing the access to subrepositories and the main repo. Things like removing a submodule is not an easy task (for a gui) and requires multiple things to take care of. 

In MsPiggit, I took a simple approach by dividing the view in two parts. One takes care of the complete repository tree and one of the current selected repository (which could be the main repo or any submodule of it). It also reads the whole tree at once, so changes are not recognized.

At least, it is not the wrong way at all.

scunz:
Frankly, after separating the MGV repository into a tree of submodules, I _do_ think that working with submodules is not anything like easy (even from command line).

antis81:
Well, when time allows it, have a look here (and following commits): https://github.com/libgit2/libgit2/commit/5f4a61aea834fe25ce1596bc9c0e0b5e563aa98b

Seems, basically it is there! I'll try to go on here.

PR #52, opened 2012-12-19
Topic/shortcuts

scunz:
I've rebased this upon `development`, leaving out the commit about the removal of ampersands.

Interestingly, git totally removed the update to the submodule, which i did not tell it to.

However, @antis81, merge it now?


antis81:
Yeah thanks. I think it's good to merge.

scunz:
So it shall be :-)

PR #59, opened 2013-03-20
Most Recently Used list for Repositories

scunz:
This patchset adds a "Recently used..." menu entry to the Repository menu.


scunz:
This depends on macgitver/libHeaven#13. This automatic issue linking seems to work only when you don't want it :-)

PR #61, opened 2013-03-23
feature: Implementation of stage area

antis81:
Currently only minor adjustments, but more to come ...

### DIFF
--- a/WorkingTree/WorkingTreeModel.cpp
+++ b/WorkingTree/WorkingTreeModel.cpp
@@ -169,36 +169,36 @@ void WorkingTreeModel::update()
     QIcon folderIcon = ip.icon( QFileIconProvider::Folder );
 
     Git::Result r;
-    Git::StatusHash sh = mRepo.statusHash( r );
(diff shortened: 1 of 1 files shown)


scunz:
@antis81, you should probably merge this intermediate state here. Without this change, the development branch won't compile...

Any thoughts on how we could track such dependencies better? I think it happened a few times now that we totally did agree that something was ready without realizing that it got dependencies.

antis81:
Upsi, ok. That's actually a hard :chestnut: and really a problem, yeah. At least the tool chain will prevent a direct merge. For the moment, we could use mail notification from jenkins.

What I'd wish for is a "task-branch", which is "automagically" generated on each dependent module. That's something I'm really missing in Git also, but currently only a dream :cloud:. Let's say, we want to setup a new task, where we can select affected modules (and add / remove some later). On these, a branch is generated on a certain commit. Afterwards we could decide to either completely or partly merge back. That would be great.

But currently, I don't have any good idea how to reliable solve this. We could setup subtasks like
- [x] depending on PR macgitver/libGitWrap#12

... but an automatic check would be far better :smile:.

PR #66, opened 2013-08-29
Improvements to repository tree

antis81:
- [x] OBSOLETE: (realized in RepoMan) - Update affected tree items when switching repositories.
- [x] idea: Display the current branch of repositories in brackets after repository name. Alternatively, this could be shown in tooltips by hovering over an item.
- [x] DEFERRED: For submodules show something similar to "git describe --tags".

### DIFF
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -145,3 +145,11 @@ QModelIndex RepoInfoModel::info2Index( RepositoryInfo* info ) const
 
     return createIndex( row, 0, info );
 }
+
+void RepoInfoModel::invalidate( RepositoryInfo *info )
+{
+    if ( !info ) return;
+
+    QModelIndex index = info2Index( info );
+    emit dataChanged( index, index );
+}
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -23,10 +23,10 @@ RepoInfoModel::RepoInfoModel()
 {
     mRepoMan = &MacGitver::repoMan();
 
-    connect( mRepoMan, SIGNAL( repositoryDeactivated(RepositoryInfo*) )
-            , this, SLOT( invalidateRepository(RepositoryInfo*) ) );
-    connect( mRepoMan, SIGNAL( repositoryActivated(RepositoryInfo*) )
-             , this, SLOT( invalidateRepository(RepositoryInfo*) ) );
+    connect( mRepoMan, SIGNAL(repositoryDeactivated(RepositoryInfo*))
+             , this, SLOT(invalidateRepository(RepositoryInfo*)) );
+    connect( mRepoMan, SIGNAL(repositoryActivated(RepositoryInfo*))
+             , this, SLOT(invalidateRepository(RepositoryInfo*)) );
 }
 
 int RepoInfoModel::rowCount( const QModelIndex& parent ) const
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -24,9 +24,9 @@ RepoInfoModel::RepoInfoModel()
     mRepoMan = &MacGitver::repoMan();
 
     connect( mRepoMan, SIGNAL( repositoryDeactivated(RepositoryInfo*) )
-            , this, SLOT( invalidate(RepositoryInfo*) ) );
+            , this, SLOT( invalidateRepository(RepositoryInfo*) ) );
     connect( mRepoMan, SIGNAL( repositoryActivated(RepositoryInfo*) )
-             , this, SLOT( invalidate(RepositoryInfo*) ) );
+             , this, SLOT( invalidateRepository(RepositoryInfo*) ) );
 }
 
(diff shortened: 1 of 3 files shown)


Repository/RepoTreeView.cpp
```diff
@@ -82,7 +82,14 @@ void RepoTreeView::onCtxActivate()
         RepositoryInfo* info = qobject_cast< RepositoryInfo* >( action->activatedBy() );
         if( info )
         {
+            RepositoryInfo* prev = MacGitver::repoMan().activeRepository();
             MacGitver::repoMan().activate( info );
+            if ( prev != info )
+            {
+                // update previous item
+                const QModelIndex prevIndex = mModel->info2Index( prev );
+                mRepos->dataChanged( prevIndex, prevIndex );
```
scunz: Are you sure, that doing this invalidation is the duty of the view? I'd prefer if this code would go into the model.

antis81: Yep. This was actually meant as a starting point for discussion. I was wondering why even the activated repo item is updated at all. I'd say this is only because the menu disappears and the items under it are "redrawn by accident"? We need to inform the model about the invalidation, but how do we do that? Or do you mean something like the following?

``` c++
mModel->invalidate( prev ); // invalidate simply would find its model index (line 90) and `emit dataChanged(prevIndex, prevIndex);` afterwards.
```

Repository/RepoTreeView.cpp
```diff
@@ -82,7 +82,13 @@ void RepoTreeView::onCtxActivate()
         RepositoryInfo* info = qobject_cast< RepositoryInfo* >( action->activatedBy() );
         if( info )
         {
+            RepositoryInfo* prev = MacGitver::repoMan().activeRepository();
             MacGitver::repoMan().activate( info );
+            if ( prev != info )
+            {
+                // update previous item
+                mModel->invalidate( prev );
```
scunz: @antis81, this change is right; but not exactly what i was thinking about: the call to `MacGitver::repoMan().activate(x)` should be complete by itself. Meaning that in `RepoInfoModel`, you should make your `invalidate()` a slot and connect it to RepoManager's `repositoryActivated` singal.

Also note, that on every such change, 2 Items must be updated. Thus, you need to teach `RepoManager` to also send a signal on de-activation of a repository and connect that to the model's `invalidate()` slot, too. Btw. I would rename it to `invalidateRepository`.

The whole reasoning behind this is, that repositories can also be activated/deactivated by other code paths - not just user interaction via ctx menu of RepoTree.
Actually, _this_ is the kind of event-system that we will sooner or later have to create (or better expand with things like: "headChanged()" etc)...

antis81:
Should we merge here in favor to get the working features in?

Btw.: It would be a nice feature for GH web-interface to create a PR out of an existing issue. Would make it easier to separate topics. Seems, this is currently possible through the API only. Or am I just blind?

Some words on the implementation:
The branch info was added in the tooltip role of an item. The decision for that was initially, that the call to `branchDisplay` returns HTML text, which is not rendered in the tree item and shown as plain text. It would be actually better to perform the HTML formatting somewhere else instead of the RepoInfo item. Anyways, it behaves like intended: Marks an active branch orange, otherwise "detached at ..." is shown.
For the submodule info it would be cool, if multiple things are shown (e.g. "git describe --tags"). Further we could add information, showing where it "was before" and where it "is now". But I think keeping it as is is good enough for now.

scunz:
> Btw.: It would be a nice feature for GH web-interface to create a PR out of an existing issue. Would make it easier to separate topics. Seems, this is currently possible through the API only. Or am I just blind?

Absolutely correct. Would be a nice addition. Maybe someone should write a mail to support@github.com. They're actually quite responsive on user feedback.

Repository/RepoInfoModel.cpp
```diff
@@ -22,6 +22,11 @@
 RepoInfoModel::RepoInfoModel()
 {
     mRepoMan = &MacGitver::repoMan();
+
+    connect( mRepoMan, SIGNAL( repositoryDeactivated(RepositoryInfo*) )
```
scunz: Between the braces of SINGAL() and SLOT() should be no spaces unless absolutely required (i.e. in `const foo` you cannot avoid the space, but most of the time the `const` is superfluous).

These are actually strings that have to be stripped off their spaces before they can be compared; which is a memory and time overhead that is simply not necessary. Also, this is the cause why namespaces must always be fully qualified in definition and usage of slots and signals...

scunz:
Well, I'd say lets get this in - and I agree on the fact that `RepositoryInfo` is not the right place for this HTML formatting.

What would you think about creating a widget that looks like the bubble in this box:
at_sacu

We could put some of them into the status bar at the bottom (which currently looks wired with the coloured branch name). And we could also use this with a text like "master +4" in the tree (grey in the usual case and in branch-colour for the active repository) 

 We'd just need the painting code in a separate method somewhere...

antis81:
Looks like a cool addition :+1:.

antis81:
Ok, I'll remove the spaces and merge this one afterwards. @scunz Think it goes well with your improvements of our RepoMan.

PR #71, opened 2013-09-12
improvements to reference view

antis81:
- [x] sort reference tree by name
- [x] introduce rename dialog for references
- [x] hide HEAD in reference tree
- [x] remove "local/remote" buttons in toolbar
- [x] add "jump to active branch" button in toolbar

### DIFF
--- a/RefsViews/Branches/BranchesView.cpp
+++ b/RefsViews/Branches/BranchesView.cpp
@@ -73,7 +73,7 @@ void BranchesView::showContextMenu(const QModelIndex &index, const QPoint &globa
     const RefItem *item = static_cast<const RefItem*>(srcIndex.internalPointer());
 
     Heaven::Menu* menu = 0;
-    if ( item && item->isContentItem() )
+    if ( item && (item->data(0, RefItem::TypeRole) == RefItem::Reference) )
     {
         menu = menuCtxMenuRefsView;
         //menu->setActivationContext( item );
--- a/RefsViews/Branches/BranchesView.cpp
+++ b/RefsViews/Branches/BranchesView.cpp
@@ -218,6 +218,30 @@ void BranchesView::onRenameRef()
     }
 }
 
+void BranchesView::onJumpToCurrentBranch()
+{
+    QModelIndexList refIndexes = mData->mModel->match(
+                mData->mModel->index(0, 0), RefItem::TypeRole, RefItem::Reference,
+                -1, Qt::MatchRecursive | Qt::MatchExactly );
+    QModelIndex foundIndex;
+    foreach ( const QModelIndex& index, refIndexes )
+    {
+        const RefBranch* item = static_cast<const RefBranch*>(index.internalPointer());
+        if ( item && item->reference().isCurrentBranch() )
+        {
+            foundIndex = index;
+            break;
+        }
+    }
+
+    if ( foundIndex.isValid() )
+    {
+        QModelIndex proxyIndex = mData->mSortProxy->mapFromSource( foundIndex );
+        mTree->scrollTo( proxyIndex );
+        mTree->selectionModel()->select( proxyIndex, QItemSelectionModel::SelectCurrent );
+    }
+}
+
 void BranchesView::actionFailed( const Git::Result& error )
 {
     if ( error ) return;
--- a/RefsViews/Branches/BranchesView.cpp
+++ b/RefsViews/Branches/BranchesView.cpp
@@ -28,6 +28,7 @@
 
 #include "RefItem.hpp"
 #include "RefsSortProxy.hpp"
+#include "RefRenameDialog.hpp"
 
 
 BranchesView::BranchesView()
@@ -199,11 +200,22 @@ void BranchesView::onRenameRef()
     Heaven::Action* action = qobject_cast< Heaven::Action* >( sender() );
     if ( !action ) return;
 
-    QModelIndex proxyIndex = mTree->currentIndex();
-    if ( !proxyIndex.isValid() ) return;
+    QModelIndex srcIndex = mData->mSortProxy->deeplyMapToSource( mTree->currentIndex() );
+    if ( !srcIndex.isValid() ) return;
 
-    // note: Resolving the proxy index is done by the QTreeView::edit().
-    mTree->edit( proxyIndex );
+    RefRenameDialog* dlg = new RefRenameDialog( this );
+    dlg->init( static_cast<RefBranch*>(srcIndex.internalPointer()) );
+
+    if ( dlg->exec() != QDialog::Accepted )
+    {
+        const Git::Result& dlgResult = dlg->gitResult();
+        if ( !dlgResult )
+        {
+            QMessageBox::warning( this, trUtf8("Failed to rename reference"),
+                                  trUtf8("Failed to rename reference:\nGit message: %1")
+                                  .arg(dlgResult.errorText()) );
+        }
+    }
 }
 
 void BranchesView::actionFailed( const Git::Result& error )
--- a/RefsViews/Branches/BranchesView.cpp
+++ b/RefsViews/Branches/BranchesView.cpp
@@ -195,8 +195,7 @@ bool BranchesView::checkRemoveRef( const Git::Reference& ref )
 void BranchesView::onRenameRef()
 {
     Heaven::Action* action = qobject_cast< Heaven::Action* >( sender() );
-    if ( !action )
-        return;
+    if ( !action ) return;
 
     QModelIndex srcIndex = mTree->currentIndex();
     if ( !srcIndex.isValid() )
@@ -223,14 +222,15 @@ void BranchesView::contextMenuEvent(QContextMenuEvent *ev)
     }
     else if( ev->reason() == QContextMenuEvent::Mouse )
     {
-        QModelIndex idx = mTree->indexAt( ev->pos() );
-        showContextMenu( idx, ev->globalPos() );
+        showContextMenu( mTree->indexAt( ev->pos() ), ev->globalPos() );
         ev->accept();
     }
 }
 
 void BranchesView::attachedToContext( Heaven::ViewContext* ctx, Heaven::ViewContextData* data )
 {
+    Q_UNUSED( ctx );
+
     BranchesViewData* myData = qobject_cast< BranchesViewData* >( data );
     Q_ASSERT( myData );
 
--- a/RefsViews/Branches/BranchesView.cpp
+++ b/RefsViews/Branches/BranchesView.cpp
@@ -27,6 +27,8 @@
 #include "libGitWrap/Reference.hpp"
 
 #include "RefItem.hpp"
+#include "RefsSortProxy.hpp"
+
 
 BranchesView::BranchesView()
     : ContextView( "Branches" )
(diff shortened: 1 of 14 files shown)


RefsViews/Branches/BranchesView.cpp
```diff
@@ -223,14 +224,15 @@ void BranchesView::contextMenuEvent(QContextMenuEvent *ev)
     }
     else if( ev->reason() == QContextMenuEvent::Mouse )
     {
-        QModelIndex idx = mTree->indexAt( ev->pos() );
-        showContextMenu( idx, ev->globalPos() );
+        showContextMenu( mTree->indexAt( ev->pos() ), ev->globalPos() );
```
scunz, line 263: No mapping required here?

scunz:
The sorting stuff looks good.

For the branch renaming: I'm not sure that it is a good idea to let the user change the branch name in place in the tree view.

P.S.: I'm actually about to create a generic Proxy Sorter with some MGV specific features inside libMacGtiverCore... But will take a few days...

antis81:
> P.S.: I'm actually about to create a generic Proxy Sorter with some MGV specific features inside libMacGtiverCore... But will take a few days...

Great! I added sorting to the RepoTree also. This is basically working, but has some flaws still. It needs cleanup within the repo model and RepoInfo signal/slot handling.

> For the branch renaming: I'm not sure that it is a good idea to let the user change the branch name in place in the tree view.

Renaming of branches should certainly be improved. I'm fine with the current solution for 0.1, if you don't mind. I agree, it was not intended, that the complete path can be changed here. Further, it should be possible to synchronize the renamed branch with with the remote repository(s). Therefore we need some kind of dialog.

RefsViews/Branches/BranchesView.cpp
antis81, line 263: The mapping is done within showContextMenu() itself. Thus, calling it here would be twice.

scunz:
> I added sorting to the RepoTree also. This is basically working, but has some flaws still. It needs cleanup within the repo model and RepoInfo signal/slot handling.

I already did that at the beginning of the week :collision: :collision: :-)

For the renames: um, well, I actually didn't think so far: But you're right this is a sleeping beast and deserves a dialog :-)

antis81:
@scunz Please review and comment.

References are now renamed in a separate dialog by using their "shorthand". Still this is just the tip of the iceberg. We need a real smart mechanism to synchronize the remote refs. But this is a bigger topic deserving a single PR like "much improved renaming of references". I'd leave that one for 0.2. For this reason I did not include any placeholders and such things.

RefsViews/RefsSortProxy.hpp
```diff
@@ -0,0 +1,41 @@
...
+public:
+    QModelIndex deeplyMapToSource( QModelIndex current ) const;
+
+private:
+    bool filterAcceptsColumn(int sourceColumn, const QModelIndex &sourceParent) const;
+    bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const;
+
+};
+
+#endif // REFS_SORT_PROXY_HPP
```
scunz, line 41: Some versions of the C++ spec don't allow these comments after a preprocessor command. :-)

RefsViews/Branches/BranchesView.cpp
```diff
@@ -195,14 +198,48 @@ bool BranchesView::checkRemoveRef( const Git::Reference& ref )
...
+
+void BranchesView::onJumpToCurrentBranch()
+{
+    QModelIndexList refIndexes = mData->mModel->match(
+                mData->mModel->index(0, 0), RefItem::TypeRole, RefItem::Reference,
+                -1, Qt::MatchRecursive | Qt::MatchExactly );
+    QModelIndex foundIndex;
+    foreach ( const QModelIndex& index, refIndexes )
+    {
+        const RefBranch* item = static_cast<const RefBranch*>(index.internalPointer());
```
scunz, line 229: Calling `internalPointer()` from somewhere outside of the model should be a no-go (from an architectural pov), shouldn't it?

RefsViews/Branches/BranchesViewData.cpp
```diff
@@ -35,12 +38,22 @@ void BranchesViewData::attachedToContext( Heaven::ViewContext* context )
...
+
+    mSortProxy->sort( 0 );
 }
 
 void BranchesViewData::detachedFromContext()
 {
     delete mModel;
     mModel = NULL;
+
+    delete mSortProxy;
```
scunz, line 52: `mSortProxy` uses the `mModel`, right? If so, shouldn't the order of deletion be reversed? I mean, in theory the destructor of the proxy model could try to access it's source model...

RefsViews/RefItem.cpp
```diff
@@ -48,11 +48,6 @@
     qDeleteAll( children );
 }
 
-bool RefItem::isContentItem() const
```
scunz, line 51: Is there a reason to remove the virtual in favour of the complex `data()` integration? I'm not against it (the speed loss is marginal) but curious.

RefsViews/RefItem.cpp
```diff
@@ -136,21 +137,21 @@ QVariant RefNameSpace::data(int col, int role) const
 
 QVariant RefBranch::data(int col, int role) const
 {
-    switch( role )
```
scunz, line 139: Why do you want to remove the switch in favour of a bunch of if's? IIRC, I have just recently changed it the other way (at least I think it was this one)...

RefsViews/RefsSortProxy.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+bool RefsSortProxy::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    QModelIndex i = sourceModel()->index( sourceRow, 0, sourceParent );
+    if ( !i.isValid() )
+        return false;
+
+    QString itemName = i.data().toString();
+    bool display = itemName.right(4) != QString::fromUtf8( "HEAD" );
```
scunz, line 60: I don't think that this is right because of several reasons:
- Why do we have to include something into the ref model in the first place just to have it removed later?
- It removes `HEAD` ref. This is partially right and partially critical:
  1. Either HEAD is a symbolic reference (the usual case) and we should display it accordingly (and do that for all symbolic refs)
  2. Or HEAD is actually some kind of tag (the detached HEAD case). How would it be accessible when not displayed? I came across the same problem in HistoryView; but I think that I haven't solved it yet.
  3. HEAD might be a symbolic ref and yet to be born (right after `git init`) We should give a special appearance to that case.
- It also removes `remotes/foo/HEAD` which is the display for the remote's default branch. We should display that at another place then.

RefsViews/RefItem.cpp
antis81, line 51: Yep, there's a reason :smile:. The `QAIM::match()` method uses QVariant comparison to find items through the `data()` method. This makes searching comfortable. When we have type switches, we don't need that anymore and from now on any items despite "Scopes" are (visually) selectable when clicking on them.

RefsViews/RefItem.cpp
antis81, line 139: The `switch` actually caused compile errors when adding another case below the last one. I guess it is due to variables declared inside a case block. After some fiddling I decided to change it to the if's. Nothing against switch case from my side. I just don't know how to handle that.

scunz, line 139: Oh, no problem then :-)

Actually, there is nothing like a "case" block. It's only the "swtich" block (scope). So all variables declared inside a switch and not nested in deeper scopes are in scope at any label. From the implementation point of view, switch is actually a "multi goto". These case marks are _really_ labels like the ones you could jump to with `goto`. Thus, the same rules apply for `goto` and `switch`.

To get around that, see this:

```
switch (x) {
case 1:
  { // <-- creates a new scope
    int i = foo();
    if (i > 100 ) return;
    break;
  } // <-- closes the scope, "i" is now no longer in scope
case 2:
    int i = foo();
    if (i > 100 ) return;
    break;
    // "i" is still in scope. If switch() ends here, everything's fine.
    // If switch doesn't end here, compiler will warn or error out because of uninitialized variable i.
case 3: // at least a warning here.
}
```

RefsViews/RefItem.cpp
scunz, line 51: Ah, ok, I see. You're using match on this for the jump to current feature.

Hm, when I read the jump-to-current code yesterday, I was thinking: why not simply match on the ref name?

RefsViews/Branches/BranchesViewData.cpp
antis81, line 52: That's a good point :smile:. I'll change the order.

scunz:
@antis81, could you merge this one in quickly but without hurry :-)

Actually, I finally started refactoring the repository manager, so this will soon clash...

antis81:
@scunz To keep things easy, feel free to merge this in favor to the refactoring. I have made only few changes yet according to our latest discussion. No need to hurry or worry :). I'll open a new PR then to get the changes right.

scunz:
This is going into the right direction anyway, but might probably be revisited when I'm done with the RepoMan refactoring. I'm going to merge this in order to not lose this changes.

scunz:
Ups, I think, I just killed your PR while you were still adding to it :8ball:

antis81:
Ah, just one minute too slow :smile: - added the check on detached head. Would you please merge.

scunz:
Good to know that git never loses anything unless you really, really, really badly tell it to... :-)

scunz:
```
-- sacu@sammi Mo Sep 23 23:25 /work/mgv/src/MacGitverModules (development u=)                                                                                                                                                     
 $ git cherry-pick 6a0f                                                                                                                                                                                                           
[development fef5e20] check if repo is valid and head is not detached before trying to highlight  the current branch                                                                                                              
 Author: Nils Fenner <nilsfenner@web.de>                                                                                                                                                                                          
 1 file changed, 15 insertions(+), 3 deletions(-)                                                                                                                                                                                 
-- sacu@sammi Mo Sep 23 23:25 /work/mgv/src/MacGitverModules (development u+1)                                                                                                                                                    
 $ git push                                                                                                                                                                                                                       
Password for 'https://scunz@github.com':                                                                                                                                                                                          
Counting objects: 9, done.                                                                                                                                                                                                        
Delta compression using up to 4 threads.                                                                                                                                                                                          
Compressing objects: 100% (5/5), done.                                                                                                                                                                                            
Writing objects: 100% (5/5), 757 bytes | 0 bytes/s, done.                                                                                                                                                                         
Total 5 (delta 4), reused 0 (delta 0)                                                                                                                                                                                             
To https://scunz@github.com/macgitver/MacGitverModules.git                                                                                                                                                                        
   b22eca6..fef5e20  development -> development
```

antis81:
Originally I was thinking about the "detached head problem" mentioned by you. Then I found a method is already implemented in GitWrap. So I just added the few lines some minutes ago and also changed to "RepoMan", which makes refactoring probably easier. No other changes in the queue.

scunz:
Yepp, that's fine.

So let's see, if I can still rebase without trouble...

scunz:
Went almost smoothly. Just one include related compile error introduced into my branch.... :-)

antis81:
Hm, yeah cool. Battery empty - cu :)

PR #78, opened 2013-10-19
finalization of welcome screen for v0.1

antis81:
- [x] final version of background image (@claus007: peaceful reminder :smile:)
- [x] transparent background image
- [x] fixed image to bottom right (even when resizing the viewport smaller than the bg image)


scunz:
@antis81, @claus007 Whatever image will finally make it, I think that it should be included in source format, too. That is probably svg here, right?

antis81:
It's actually Blender format (a .blend text file). Maybe it makes sense to add a submodule containing that :question:.

scunz:
No submodule, just put the image source alongside the final image, so it stays reproducible.

scunz:
:+1: :page_facing_up: :100:

PR #82, opened 2013-11-02
follow gitwrap ref_stuff branch

antis81:
This PR is following the current development in GitWrap and fixes stuff. Should be merged as soon as GitWrap stuff is ready.

### DIFF
--- a/History/HistoryBuilder.cpp
+++ b/History/HistoryBuilder.cpp
@@ -125,7 +125,7 @@ void HistoryBuilder::start()
 {
     QVector< Git::ObjectId >	commits;
     Git::ObjectId				currentSHA1;
-    Git::ObjectCommit			curCommit;
+    Git::Commit                 curCommit;
     Git::Result					r;
     qint64						dur;
     double						avg;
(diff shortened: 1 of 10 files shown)


scunz:
Thanks for the Header think. I'm actually not yet compiling MGV against new gitWrap, yet. But when I did (before the header change), there were more than 2 deprecation warnings. Did you intend to catch them all or is this just a random fix?

antis81:
> Did you intend to catch them all or is this just a random fix?

It is not a random fix. We can use it to follow the new GitWrap features and merge to development. I just wasn't sure, if you're working already on swiping out deprecated warnings. At this place, there's no warning - which I saw to late should have been there. Some warnings need to be fixed in MGVCore.

scunz:
Fine. I did not pay too much attention to the deprecation warnings (After all, I created these in order to do it not _now_ this time). I was just wondering, because I see roughly 100 warnings - and you're fixing just 2.

antis81:
> ...I see roughly 100 warnings - and you're fixing just 2.

:) - I also see the warnings, but these 2 gave compile errors. Added the deprecated function now locally, I'll add it as another PR to GitWrap, letting you work on :walking:. But you're right, it is better to do the real fix after merging the big RM branch to make it bright and shiny again :star2:.

scunz:
And this one, too...

antis81:
Thanks.

PR #85, opened 2013-12-08
Working Tree: Restore filter behavior from before using ProxyFilters

scunz:
A thing that bothered me for quiete a while now: When I filter away unchanged entries in the WT, all directories are still shown.

This is regression was introduced with the moving to the proxy model for filtering. The actual cause is that we don't recursively check if the filter rules apply to all children. But this was hidden by a "workaround" bugfix (DirItems didn't report status, so they were just always accepted).

This commit fixes both issues and restores the behaviour before the change to the filter proxy...

### DIFF
--- a/WorkingTree/WorkingTreeDirItem.cpp
+++ b/WorkingTree/WorkingTreeDirItem.cpp
@@ -58,6 +58,9 @@ QVariant WorkingTreeDirItem::data( int column, int role ) const
         }
         break;
 
+    case StatusRole:
+        return 0;
+
     default:
         break;
     }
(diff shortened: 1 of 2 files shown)


scunz:
@antis81, please confirm that this is right and actually a "sane" behaviour (Just want to make sure you didn't implement it that way be intention, in which case we should discuss it a bit :smile:)

OTOH, feel free, to merge, if this change is okay.

scunz:
Thanks :)

antis81:
Yep, the intention was to show directories only in the index, when they really changed / are not empty. But at that time we set focus on other topics. Looks good, thanks. Another nice feature what I have in mind for months now is switching between a list and tree view. But that's actually nice to have.

scunz:
One of the next steps will be to tear the `WorkingTreeView` apart and create a new `Commit`-Module/View/Mode. Will think about that, too.

PR #86, opened 2013-12-29
feature: commit changes to repository

antis81:
- [x] add commit dialog to WorkTree/Index view
- [x] create commit and add to repository (implementation in GitWrap needed)

### DIFF
--- a/WorkingTree/CMakeLists.txt
+++ b/WorkingTree/CMakeLists.txt
@@ -10,6 +10,8 @@ INCLUDE_DIRECTORIES(
 
 SET( SRC_FILES
 
+    CommitDialog.cpp
+
     IndexWidget.cpp
     IndexTreeItemView.cpp
     WorkingTreeModule.cpp
@@ -23,6 +25,8 @@ SET( SRC_FILES
 
 SET( HDR_FILES
 
+    CommitDialog.h
+
     IndexWidget.h
     IndexTreeItemView.h
     WorkingTreeModule.h
@@ -34,6 +38,11 @@ SET( HDR_FILES
     WorkingTreeFilterModel.h
 )
 
+SET( UI_FILES
+
+    CommitDialog.ui
+)
+
 SET( HID_FILES
 
     WorkingTreeActions.hid
@@ -41,6 +50,7 @@ SET( HID_FILES
     StageViewCtxMenu.hid
 )
 
+QT_UIC( UIC_FILES ${UI_FILES} )
 QT_MOC( MOC_FILES ${HDR_FILES} )
 HIC( HIC_FILES ${HID_FILES} )
 
@@ -50,6 +60,7 @@ ADD_MGV_MODULE(
     ${SRC_FILES}
     ${HDR_FILES}
     ${MOC_FILES}
+    ${UIC_FILES} ${UI_FILES}
     ${HIC_FILES} ${HID_FILES}
 )
 
--- a/WorkingTree/CommitDialog.cpp
+++ b/WorkingTree/CommitDialog.cpp
@@ -19,8 +19,11 @@
 #include "CommitDialog.h"
 #include "ui_CommitDialog.h"
 
+#include "libGitWrap/Result.hpp"
+
 #include <QPlainTextEdit>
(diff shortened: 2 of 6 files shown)


scunz:
@antis81, actually, when I first read this, I wanted to immediately answer that this doesn't fit the concept very much and there're lots of small things inside that I don't like. But I've been thinking about that since - and I'm not going to tell this story.
Well, meanwhile I indeed think that it is absolutely of no importance _how_ we reach the goal. Actually, I changed my mind from a total :-1: to an absolutely :+1: :-)
So, I'm just going to ask for a few things for the longer term:
- Please ensure that the commit code can be easily moved out of the WorkingTree Module (actually, this changeset is a right step into that direction already, but the code to actually do the commit should be moved to the new class).
- Please ensure that you name things logically: If it is a Dialog, call it a Dialog and use it as a Dialog - If you don't want to use it as a Dialog, then don't call it one :)
- But the only half way serious point left is: Let's not try to out-smart the user, but stick to what git provides in the first place. I'm talking about the "Empty commit message"-thing here. Git itself actually allows that (but needs to be convinced that it is a good idea via command line parameter), so we might come across a commit that has no message during an `--amend` or edit operation...

antis81:
@scunz the questions you ask are absolutely good ones :)

:exclamation: First of all, this feature is essential, because it is one making MGV actually usable. This is also the reason I startet this PR.
- Named it Dialog because it is something inbetween a dialog and a widget ... I'll rename it to widget instead.
- The decision of code location is also essential -> I'll move everything inside the CommitWidget class.
- Amending is another story. I take the GitX commit dialog as an excellent inspiration for that (it visualizes the index and WT structure during amend). Grouping commit functionality into a separate module is surely a good idea. As said above I'll move code into the same class.
- Last but not least I'll change the empty check as well. At that point it was actually just meant as a place holder, not a final solution. I'm looking forward to the commit functionality in GitWrap itself ^^. Btw. lg2 v0.20 (and thus our current development branch) doesn't compile on mac - same bug as on Sun OS (think it was some strlen function is not supported). This should be fixed before.

scunz:
> Btw. lg2 v0.20 (and thus our current development branch) doesn't compile on mac - same bug as on Sun OS (think it was some strlen function is not supported). This should be fixed before.

I am compiling it just fine on Mac OS X (10.9.1/XCode 5) must be something with Mac versions.

scunz:
> First of all, this feature is essential, because it is one making MGV actually usable.

Partially agree: Yep, it's true - but we're a good viewer already.

antis81:
> I am compiling it just fine on Mac OS X (10.9.1/XCode 5) must be something with Mac versions.

fixed in lg2 development branch :smile:

EDIT: @scunz Could you please push lg2 changes manually to mgv/lg2 fork? Thanks in advance.
EDIT2: Sorry for being impacience - I did it :)

antis81:
Rebased on `development` to take advantage of the "purified skies" changes :100:

scunz:
> EDIT: @scunz Could you please push lg2 changes manually to mgv/lg2 fork? Thanks in advance.
> EDIT2: Sorry for being impacience - I did it :)

Apparently The @-thingie in combination with edits doesn't work. So, actually I missed this one. But, don't worry too much: Will soon be done automagically again... Think that I will write a mail tomorrow. But I just realised that the next week will be "stressful" again. So, I hope to finish as much as possible tomorrow.

antis81:
> Apparently The @-thingie in combination with edits doesn't work. So, actually I missed this one. But, don't worry too much: Will soon be done automagically again... Think that I will write a mail tomorrow. But I just realised that the next week will be "stressful" again. So, I hope to finish as much as possible tomorrow.

I've done that already :smile: - sorry, I forgot to write before.

scunz:
> I've done that already :smile: - sorry, I forgot to write before.

Yep, I think I saw you doing that - and I wanted to ask for creation of a tag for it. But actually, it is likely that we'll switch back to development branch of libgit2 again, if the infrastructure is ready again...

antis81:
Think we can merge this after the PR in libGitWrap

PR #89, opened 2014-03-15
Context menu for history view

antis81:
Let's add some action to the context menu, giving it abilities to interact with the history.
- [x] add context menu

Give it an "easy start" by adding cool `libGitWrap` stuff, that is almost ready to use and just needs some finalization there.

So here comes the first actions:
- [x] checkout a commit in history
  If a local branch is pointing here, we should ask the user, if he wants to checkout and activate it. We could also create a new branch, but this workflow is provided by the following action either.
- [x] create a (local) branch on a selected history item
  Later, we could think of creating remote branches as well, but that is pretty advanced stuff. In the first run we leave that out and keep the workflow simple.

### DIFF
--- a/History/CMakeLists.txt
+++ b/History/CMakeLists.txt
@@ -46,6 +46,7 @@ SET(UI_FILES
 SET(HID_FILES
 
     HistoryViewActions.hid
+    HistoryCtxMenu.hid
 )
 
 SET(RCC_FILES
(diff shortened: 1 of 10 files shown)


antis81:
We cannot create tags yet. The rest works fine. Checkout operation has to be fixed though.

So ... should we merge as well?

antis81:
Even if not fully completed, this adds more value to MGV. I'll merge here and improve later on :smile:.

PR #90, opened 2014-04-06
display author/committer mail in separate columns

antis81:
Just a small one. This makes it possible to hide the e-mail column - e.g. when doing screenshots :smile: :bulb:. Currently it is not possible to hide certain columns though - but here's one use case for this.

### DIFF
--- a/History/HistoryModel.cpp
+++ b/History/HistoryModel.cpp
@@ -175,10 +175,10 @@ QVariant HistoryModel::headerData( int section, Qt::Orientation orientation, int
     case colGraph:          return trUtf8( "Graph" );
     case colMessage:        return trUtf8( "Message" );
     case colAuthor:         return trUtf8( "Author" );
-    case colAuthorMail:     return trUtf8( "Author e-mail" );
+    case colAuthorMail:     return trUtf8( "Author mail" );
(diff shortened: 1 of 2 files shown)


antis81:
I bet we are ready to merge this.

History/HistoryModel.cpp
```diff
@@ -171,8 +175,10 @@ QVariant HistoryModel::headerData( int section, Qt::Orientation orientation, int
     case colGraph:          return trUtf8( "Graph" );
     case colMessage:        return trUtf8( "Message" );
     case colAuthor:         return trUtf8( "Author" );
+    case colAuthorMail:     return trUtf8( "Author e-mail" );
```
scunz: Should remove the dash here, probably. (same for committer below)

scunz:
Other than that, this change is nice :)

History/HistoryModel.cpp
antis81: Hm, `e-mail` just looks nicer than `email` to me. Maybe we just should write `mail` then?

scunz: Yeah, indeed just `mail` would probably fit even better.

PR #91, opened 2014-05-03
show icons in RepoTree

antis81:
Let's add a little more color :rainbow: to the otherwise boring repository view.

### DIFF
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -17,6 +17,9 @@
 #include "libMacGitverCore/App/MacGitver.hpp"
 #include "libMacGitverCore/RepoMan/RepoMan.hpp"
 
+#include "libHeavenIcons/IconRef.hpp"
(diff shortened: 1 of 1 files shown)


antis81:
together with the new icons looks like this:
mgv_repo_history_view

scunz:
@antis81, this is really nice, but there's a big big **BUT**:

Can we somehow optimise the height ratio between the icons and the texts, ideally approaching a value of 1?

antis81:
@scunz, you mean like this ...
repotree_24pt

The first sample uses a font size of 24pt (pixelSize() again returns -1 - strange...). The second sample uses a font size of 11pt. I also introduced a minimum icon size of fixed 16px. This is what we see here:
repotree_11pt

Ok?

scunz:
Yeah, exactly. But note, that it has to match whatever the user chooses to setup... Would be funny to force the "Fontsize" to be a compile time setting :)

PR #92, opened 2014-05-04
set icon size according to the font size in repo tree

antis81:
Relates to PR #91. Unfortunately I cannot reopen an already merged PR and set it on top - sigh. Anyways, a short review would be nice.

### DIFF
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "libMacGitverCore/App/MacGitver.hpp"
+#include "libMacGitverCore/Config/Config.h"
 #include "libMacGitverCore/RepoMan/RepoMan.hpp"
 
 #include "libHeavenIcons/IconRef.hpp"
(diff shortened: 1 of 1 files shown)


Repository/RepoInfoModel.cpp
```diff
@@ -66,7 +67,13 @@ QVariant RepoInfoModel::data( const QModelIndex& index, int role ) const
             return info->displayAlias();
 
         case Qt::DecorationRole:
-            return info->icon().icon().pixmap();
+            {
+                Heaven::IconRef icon = info->icon();
+                QFont f;
```
scunz, line 72: Can you make a private function that returns the right font? We should use the same font for determining the height as we use in the `Qt::FontRole` case below. It's just cosmetic, but one day it may make a difference...

Repository/RepoInfoModel.cpp
```diff
@@ -66,7 +67,13 @@ QVariant RepoInfoModel::data( const QModelIndex& index, int role ) const
...
+                int size = (f.pixelSize() <= 0) ? f.pointSize() : f.pixelSize();
+                icon.setSize( qMax(size + 2, 16) );
```
scunz, line 74: This feels wrong, when it should be _litterally_ alright actually.

The problem is: IconRef is an _explicitly_ shared class, not a Qt-ish _implicit_ shared class. This means, you're actually modifying the `Heaven::IconRef` instance that is stored inside the `RM::Repo`. I don't think this is the intention (but probably won't harm - it just isn't right). However, if someday we're relying on the `RM:Repo::icon()` to be exactly what we set it to... Well. It will probably take ages to figure why the icon size is wrong...

So... We could either make the `IconRef` class implicitly shared or add a `clone()` method to it.

antis81, line 74: I think `clone()` is fine then. Thanks for pointing that out.

scunz:
Looking fine now, just got it in :)

PR #94, opened 2014-05-11
coloring ref. labels within history

antis81:
This looks beatiful in my eyes :). Colors taken from GitX. The color for the current branch is the same everywhere in MGV now, which is a nice thing.

These colors could be customizable within the settings - but that's a topic for v0.2. For now let's take fixed colors and make a redmine ticket out of the issue :).

This is how it looks:
mgv-history-colors

@claus007 FYI :smile:

### DIFF
--- a/History/HistoryListDelegate.cpp
+++ b/History/HistoryListDelegate.cpp
@@ -290,8 +290,6 @@ void HistoryListDelegate::paintMessage( QPainter* p, const QStyleOptionViewItem&
             p->setFont( opt.font );
 
             QRect refRect( r.left(), r.top() + 1, w, r.height() - 3 );
-            QPainterPath refPath;
-            refPath.addRoundRect( refRect, 20 );
 
             QColor back = QColor( 0xD9D9D9 );
             if( ref.mIsTag )
@@ -308,13 +306,16 @@ void HistoryListDelegate::paintMessage( QPainter* p, const QStyleOptionViewItem&
                     back = QColor( 0x9BFD51 );
             }
 
-            p->fillPath( refPath, back );
-            QPen blackPen(Qt::black);
-            blackPen.setWidth(0);
-            p->setPen(blackPen);
-            p->drawPath( refPath );
+            QLinearGradient gradient( refRect.left(), 0, refRect.right(), 0 );
+            gradient.setColorAt( 0.0, back.lighter(125) );
+            gradient.setColorAt( 0.33, back );
+
+            p->setBrush( gradient );
+            p->setPen( QPen(Qt::transparent) );
+            p->drawRoundedRect( refRect, 2.5, 2.5 );
 
             refRect.adjust( 2, 0, -2, 0 );
+            p->setPen( QPen(Qt::black) );
             p->drawText( refRect, Qt::AlignCenter, ref.mRefName );
 
             r.setLeft( r.left() + w + 3 );
--- a/History/HistoryListDelegate.cpp
+++ b/History/HistoryListDelegate.cpp
@@ -293,7 +293,7 @@ void HistoryListDelegate::paintMessage( QPainter* p, const QStyleOptionViewItem&
             QPainterPath refPath;
             refPath.addRoundRect( refRect, 20 );
 
-            QColor back = Qt::white;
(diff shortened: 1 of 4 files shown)


scunz:
Heya @antis81, please go ahead with this change :)

But when you're at this, can you please take care of a few other things, too:
-    Consider to change the color of the most recent STASH to something more beautiful:
  mgv_stash
-   Reset the font to non-bold after what ever brach-box has been drawn.
-   Could you give them a slight gradient (and play with removing the black border, too). I am obviously no clone of Picasso, but I could imagine that this would fit nicer with the other layout elements.
-    Can you give the bulbs on the graph a similar colouring schema (i.e. use more vivid colours for them, too)

antis81:
After lots ... and lots of fiddeling around with several gradients and borders, I found two nice looking patterns, which seem pretty reasonable to me.

The first one is with a lighter gradient and a slight border:
labels-grad-border

The second is the same gradient without border:
labels-grad-noborder

Phew ... this one somewhat gave me nuts :nut_and_bolt: :nut_and_bolt: and I learned, that I have to give exact coordinates for the gradient's start and end points. They are not translated to the rect and produce strange results otherwise :8ball:.

Also fixed some other things (see code :smile:) ...

scunz:
This is looking really great! It kind of feels like these things have really been marked with a text marker, which is pretty cool actually. Maybe, though, the gradient should fade out at the end, too?

scunz:
Another question: Should we add a center-ellipsis if the text gets too long? Like in `origin/.../foo-bar-baz`?

antis81:
Currently we have right ellipsis. Center ellipsis would also be ok. I'll change it and commit ...

antis81:
Hmz, as already said I fiddled around a lot with that and I quite find "fade in" looks far better than "fade out" or both combined. But, see for yourself:
labels-grad-lr-elidemiddle

The "fade out" might have the effect, that the label text might look like part of the message.

Also tried the width restriction (see screenshot) for the gradient. Plus, it samples the "elide middle" variant. Had to reduce min. width to 30px to make a visible difference. What Do you think? Commit or regret?

antis81:
Oops, I got you wrong about the ellipsis ... Introducing ellipsis in the label text is a good idea. But ... we'd also need a solution to show the full path again. Thinking of 10 people working on their `origin/peter/.../feature`, `origin/paul/.../feature`, etc. branch. So for example, hovering the mouse over a label could show the tooltip with the full branch name.

Edit:
Might be better a feature for 0.2 ^^

antis81:
suggestion for graph colors:
vivid-graph

Ok?

scunz:
The graph colours themselves are obviously better. However, they should not repeat inside the array, since that is done automatically (by cycling the array via modulus). The two greens and the two purple look very similar.

> The "fade out" might have the effect, that the label text might look like part of the message.

In the picture, I can see that happen only on the remote branches. So, maybe you should not use the lighter() function on the colour, but actually specify two different colours. (i.e. make the gradient colour on the remote-branches more like `lighter(10)` - while the others are okay actually).

> Also tried the width restriction (see screenshot) for the gradient. Had to reduce min. width to 30px to make a visible difference.

I didn't measure how long these things actually are :) But 30px sounds reasonable. Remember that it might look very differently with other font settings (will post a screen with `Helvetica Neue`)

> Plus, it samples the "elide middle" variant.

Actually, looking esp. at the first line, eliding the commit topic in the middle rather than at the end is a thing that we should probably do anyway... :)

But I would prefer to elide multi-segmented remote branches (and actually only those) before we elide the commit message. I'll try to come up with some logic for that. The wasted reference-name screen estate has bothered me a few times already.... So, it is really annoying when these take more space and the essential information becomes unavailable.
The thing is: esp. on multi-segmented references, the user is most likely very aware of what they look like (except for your very valid example, of course).

scunz:
Looks like that, here: :)
mgv_history_lanes
And:
mgv_referencebadges

Could it be that this light green is so light, that col.lighter() returns the same color - or at least a very very similar one?

antis81:
I'd like to merge ...

scunz:
Yes, Please! Erm, No, I was faster :)

PR #95, opened 2014-05-11
provide a little more info in the repo's tooltip

antis81:
Here's how it looks:
mgv-sm-info

@claus007 FYI

### DIFF
--- a/Repository/RepoInfoModel.cpp
+++ b/Repository/RepoInfoModel.cpp
@@ -79,7 +79,15 @@ QVariant RepoInfoModel::data( const QModelIndex& index, int role ) const
             return info->isActive();
 
         case Qt::ToolTipRole:
-            return trUtf8( "Branch: %1" ).arg( info->branchDisplay() );
+            return QLatin1String("<html><head><style type=\"text/css\">"
+                                 ".h1 { margin: 0px; font-size: large; color: gray; }"
+                                 ".row { margin: 0px; }"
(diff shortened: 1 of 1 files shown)


scunz:
Yeah. Nice one.

However, once more this shows: MGV should globally define a number of _relevant digits for a SHA1_ (Editable in the _preference_ dialog) and stick to that setting where ever we output SHA1s.

I.e. With a default setting of 7 digits, the text above would look like: `Detached @ a1f280a`.

Note that I'd also remove the "Branch: " label in the detached-case, since it's very confusing. This is no branch in the classical sense. What we actually want to say would look like one of those:
-    `HEAD is pointing to the tip of [refs/branches/]foo branch.`
-    `HEAD is pointing to the tip of [refs/tags/]boozoo tag.`
-    `HEAD is detached from ABCDEF0, which is 12 commits ahead of [refs/tags/]V1235.`
-    `HEAD is pointing to an unborn branch.`
-    `HEAD is pointing to the tip of [refs/branches/]master, which is 5 commits ahead and 12 commits behind it's upstream [refs/origin/branches/]master.`

Contents in `[]` just added as an aid...

scunz:
Guess this one was swallowed by time? Do you want to merge it?

antis81:
Uhm, yeah. Not a big one and some info missing. But Merge is good here I think to get it going :smile:.

PR #97, opened 2015-01-21
Update the models on object changes

antis81:
# Description

This is the next big step to bring the models in sync' with changes in the Git repository. Those changes can origin from either inside MacGitver or a random application (like i.e. Git CLI).

When a Git repository changes, MacGitver must update its data models to work on the correct data.

For example a new commit, branch, tag, ... was created. MacGtiver has to react and update its data models, which in turn trigger an update on the bound views.

All changes are detected by a central instance - the repository manager `MacGitver::repoMan()` - which calls signals on a "detected change".
# Here is a list of possible events for orientation.
## Module "Repository"
- [x] repo opened
- [x] repo closed
- [x] repo activated
- [x] submodule activated
- [x] ~~HEAD of any "opened" repo changed~~
  - We don't need that!
## Module "History"
- [x] update inline references
  - [x] ref created
  - [x] ref deleted -> see `libMacGitverCore` [PR 22 there](https://github.com/macgitver/libMacGitverCore/pull/22)
  - [x] ~~ref renamed~~ -> This is not possible, because the reference name is the "lookup key". If it changes, it would generate a "refDeleted" and a "refCreated" event.
  - [x] ref moved (the "reference target" changed)
  - [x] HEAD changed (or any symbolic reference) -> :exclamation: See [here](https://redmine.macgitver.org/issues/237)
- [x] ~~update commit list~~ -> Needs to be realized in new PR.
## Module "RefsViews"
- [x] ref created
- [x] ref deleted
- ~~ref renamed~~ -> see  above in "Module "History""  
- [x] ref moved (detects also a changed HEAD)
## Status-View
- [x] ref moved (detects also a changed HEAD)
  - see `libMacGitverCore` [PR 22 there](https://github.com/macgitver/libMacGitverCore/pull/22)
### WorkTree and Index/Stage View

Updating the work tree and the index has to go in another PR. Currently MGV does not react to changes in the working tree or index.
# Migration of the "Reference View"

Please read dicussion [in Redmine](https://redmine.macgitver.org/issues/213)

The reference view should be migrated to better integrate `RM::RepoMan`. Changing this now would make this PR unnecessarily complex.
# Other changes / improvements
## Reference View
- :+1: The reference tree is now expanded in initial state.
- :+1: The reference tree will no more collapse unexpectedly after changes.
## History View
- :+1: Inline References are now sorted alphanumerically by type (fixed priorities) - yeah!
  - If you're interested look here: https://github.com/macgitver/MacGitverModules/commit/7b9d2ced67cf197fbeb5205aafbf2e79ddbe00b1#diff-53684d014451b869081860fdde26e03bR89
- :+1: Fixed recognition of status changes in inline references. 
- :+1: The "default" diff behaviour is now "diff to all parents". By default, the diff view shows the changes in the selected commit.
  - Technically, there is no default behaviour. "All parents" is just the first entry in the list. Setting a default behaviour is a candidate for the history settings.
- :+1: Display a detached HEAD

### DIFF
--- a/History/HistoryDiff.cpp
+++ b/History/HistoryDiff.cpp
@@ -52,6 +52,8 @@ HistoryDiff::HistoryDiff()
     mToolbar = new QToolBar;
 
     mDiffTo = new QComboBox;
+    mDiffTo->addItem( trUtf8( "All parents" ), DTT_AllParents );
+    mDiffTo->addItem( trUtf8( "Parent" ), DTT_Parent );
     mDiffTo->addItem( trUtf8( "Workingtree" ), DTT_WT );
     mDiffTo->addItem( trUtf8( "Index" ), DTT_Index );
     mDiffTo->addItem( trUtf8( "Workingtree + Index" ), DTT_WT_and_Index );
@@ -59,8 +61,6 @@ HistoryDiff::HistoryDiff()
     //mDiffTo->addItem( trUtf8( "SHA-1" ), DTT_AnySHA1 );
     mDiffTo->addItem( trUtf8( "Branch" ), DTT_Branch );
     mDiffTo->addItem( trUtf8( "Tag" ), DTT_Tag );
-    mDiffTo->addItem( trUtf8( "All parents" ), DTT_AllParents );
-    mDiffTo->addItem( trUtf8( "Parent" ), DTT_Parent );
 
     mToolbar->addWidget( new QLabel( trUtf8( "Diff to:" ) ) );
     mToolbar->addWidget( mDiffTo );
--- a/History/HistoryEntry.h
+++ b/History/HistoryEntry.h
@@ -70,7 +70,6 @@ struct HistoryInlineRef
         }
 
         return  mRefName    == other.mRefName       &&
-                mRefName    == other.mRefName       &&
                 mIsBranch   == other.mIsBranch      &&
                 mIsCurrent  == other.mIsCurrent     &&
                 mIsRemote   == other.mIsRemote      &&
--- a/History/HistoryEntry.h
+++ b/History/HistoryEntry.h
@@ -81,23 +81,18 @@ struct HistoryInlineRef
     inline bool operator !=(const HistoryInlineRef& other) const {
         return !( *this == other );
     }
-};
 
-typedef QList< HistoryInlineRef > HistoryInlineRefs;
-
-struct HistoryInlineRef_LessThan
-{
-    inline bool detachedHEAD(const HistoryInlineRef& ref) const
+    inline bool detachedHEAD() const
     {
-        return ref.mIsDetached && !(ref.mIsTag || ref.mIsBranch || ref.mIsStash);
+        return mIsDetached && !(mIsTag || mIsBranch || mIsStash);
     }
 
-    inline bool nameLessThan(const HistoryInlineRef& a, const HistoryInlineRef& b) const
+    inline bool nameLessThan(const HistoryInlineRef& other) const
     {
-        return a.mRefName < b.mRefName;
+        return mRefName < other.mRefName;
     }
 
-    bool operator ()(const HistoryInlineRef& a, const HistoryInlineRef& b) const
+    bool operator <(const HistoryInlineRef& other) const
     {
         // sort order:
         // - tag
@@ -108,38 +103,40 @@ struct HistoryInlineRef_LessThan
         // - stash
         // - Everything else is sorted by mRefName
 
-        if ( a.mIsTag ) {
+        if ( mIsTag ) {
             // tags have highest priority
-            return b.mIsTag ? nameLessThan(a, b) : true;
+            return other.mIsTag ? nameLessThan(other) : true;
         }
 
-        if ( detachedHEAD( a ) )
+        if ( detachedHEAD() )
         {
             // a detached HEAD comes right after tag ? -> There can only be one!
-            return !b.mIsTag;
+            return !other.mIsTag;
         }
 
-        if ( a.mIsBranch )
+        if ( mIsBranch )
         {
-            if ( b.mIsTag || b.mIsCurrent || detachedHEAD( b ) )
+            if ( other.mIsTag || other.mIsCurrent || other.detachedHEAD() )
                 return false;
 
-            if ( a.mIsCurrent ) {
+            if ( mIsCurrent ) {
                 // a is the current branch
                 return true;
             }
 
-            return ( a.mIsRemote == b.mIsRemote ) ? nameLessThan( a, b ) : !a.mIsRemote;
+            return ( mIsRemote == other.mIsRemote ) ? nameLessThan( other ) : !mIsRemote;
         }
 
-        if ( a.mIsStash ) {
-            return b.mIsStash ? nameLessThan(a, b) : !(b.mIsTag || b.mIsBranch || detachedHEAD( b ));
+        if ( mIsStash ) {
+            return other.mIsStash ? nameLessThan( other )
+                                  : !(other.mIsTag || other.mIsBranch || other.detachedHEAD());
         }
 
-        return !(b.mIsTag || b.mIsBranch || b.mIsStash || detachedHEAD( b )) ? nameLessThan( a, b ) : false;
+        return !(other.mIsTag || other.mIsBranch || other.mIsStash || other.detachedHEAD()) ? nameLessThan( other ) : false;
     }
 };
 
+typedef QList< HistoryInlineRef > HistoryInlineRefs;
 
 class HistoryEntry
 {
--- a/History/HistoryEntry.h
+++ b/History/HistoryEntry.h
@@ -61,9 +61,86 @@ struct HistoryInlineRef
     bool        mIsRemote   : 1;
     bool        mIsTag      : 1;
     bool        mIsStash    : 1;
+    bool        mIsDetached : 1;
+
+    bool operator ==(const HistoryInlineRef& other) const
+    {
+        if ( this == &other ) {
+            return true;
+        }
+
+        return  mRefName    == other.mRefName       &&
+                mRefName    == other.mRefName       &&
+                mIsBranch   == other.mIsBranch      &&
+                mIsCurrent  == other.mIsCurrent     &&
+                mIsRemote   == other.mIsRemote      &&
+                mIsTag      == other.mIsTag         &&
+                mIsStash    == other.mIsStash       &&
+                mIsDetached == other.mIsDetached    ;
+    }
+
+    inline bool operator !=(const HistoryInlineRef& other) const {
+        return !( *this == other );
+    }
 };
 
-typedef QVector< HistoryInlineRef > HistoryInlineRefs;
+typedef QList< HistoryInlineRef > HistoryInlineRefs;
+
+struct HistoryInlineRef_LessThan
+{
+    inline bool detachedHEAD(const HistoryInlineRef& ref) const
+    {
+        return ref.mIsDetached && !(ref.mIsTag || ref.mIsBranch || ref.mIsStash);
+    }
+
+    inline bool nameLessThan(const HistoryInlineRef& a, const HistoryInlineRef& b) const
+    {
+        return a.mRefName < b.mRefName;
+    }
+
+    bool operator ()(const HistoryInlineRef& a, const HistoryInlineRef& b) const
+    {
+        // sort order:
+        // - tag
+        // - detached HEAD
+        // - current branch
+        // - local branch
+        // - remote branch
+        // - stash
+        // - Everything else is sorted by mRefName
+
+        if ( a.mIsTag ) {
+            // tags have highest priority
+            return b.mIsTag ? nameLessThan(a, b) : true;
+        }
+
+        if ( detachedHEAD( a ) )
+        {
+            // a detached HEAD comes right after tag ? -> There can only be one!
+            return !b.mIsTag;
+        }
+
+        if ( a.mIsBranch )
+        {
+            if ( b.mIsTag || b.mIsCurrent || detachedHEAD( b ) )
+                return false;
+
+            if ( a.mIsCurrent ) {
+                // a is the current branch
+                return true;
+            }
+
+            return ( a.mIsRemote == b.mIsRemote ) ? nameLessThan( a, b ) : !a.mIsRemote;
+        }
+
+        if ( a.mIsStash ) {
+            return b.mIsStash ? nameLessThan(a, b) : !(b.mIsTag || b.mIsBranch || detachedHEAD( b ));
+        }
+
+        return !(b.mIsTag || b.mIsBranch || b.mIsStash || detachedHEAD( b )) ? nameLessThan( a, b ) : false;
+    }
+};
+
 
 class HistoryEntry
 {
--- a/History/HistoryModel.cpp
+++ b/History/HistoryModel.cpp
@@ -21,6 +21,8 @@
 #include <QElapsedTimer>
 
 #include "libMacGitverCore/App/MacGitver.hpp"
+#include "libMacGitverCore/RepoMan/RepoMan.hpp"
+#include "libMacGitverCore/RepoMan/Ref.hpp"
 
 #include "libGitWrap/Reference.hpp"
 
@@ -37,6 +39,8 @@ HistoryModel::HistoryModel( const Git::Repository& repo, QObject* parent )
 
     mDisplays = 0;
 
+    RM::RepoMan &rm = MacGitver::repoMan();
+    connect( &rm, SIGNAL(refCreated(RM::Repo*,RM::Ref*)), this, SLOT(onRefCreated(RM::Repo*,RM::Ref*)) );
     Q_ASSERT( mRepo.isValid() );
 }
 
@@ -221,6 +225,11 @@ void HistoryModel::beforeAppend()
 void HistoryModel::afterAppend()
 {
     endInsertRows();
+}
+
+void HistoryModel::onRefCreated(RM::Repo* repo, RM::Ref* ref)
+{
+    scanInlineReferences();
 }
 
 void HistoryModel::ensurePopulated( int row )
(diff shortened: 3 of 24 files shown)


antis81:
@scunz Would be nice, if you could verify this change bfd7051.

antis81:
@scunz :+1: Some good news:
In the current implementation state, the `HistoryModel` and `ReferenceModel` both are aware of "reference creation". Thus, independent of the source (inside/outisde MGV), the views will get updated correctly based on the RepoMan events.

When there's light :bulb:, there's also some shadow:
Seems like the `RM::EventsInterface` needs extensions. Currently I don't get a signal when ...
1. ... HEAD was set (i.e. the active branch changed).
2. ... a reference was removed
3. ... an object was created (i.e. a new commit)

:question: Do we need definitions for them? Or is there something I overlooked here?

antis81:
I'm somewhat excited of what is available in this PR :+1:. It brings MGV a big step closer to a state, that's "publicly presentable". Well, some heavy weight bugs are still waiting to be fixed (especially memory consumption in the DiffViewer) ... but that's the next big story.

After completing some important things, I'm going to merge soon.

@scunz If you find any time, please help with the "refresh" topic. The most important parts happen in libMacGitverCore. I contribute my local work there as well - you shall see. At least, now we have real consumers for our events.

@BugSniffer If you want to know why I talk confusing stuff sometimes - just look here :smile:.

History/HistoryEntry.h
```diff
@@ -61,9 +61,86 @@ struct HistoryInlineRef
...
+                mIsStash    == other.mIsStash       &&
+                mIsDetached == other.mIsDetached    ;
+    }
+
+    inline bool operator !=(const HistoryInlineRef& other) const {
+        return !( *this == other );
+    }
+};
+
+typedef QList< HistoryInlineRef > HistoryInlineRefs;
```
scunz: Why is this a list now, instead of a vector?

History/HistoryEntry.h
```diff
@@ -61,9 +61,86 @@ struct HistoryInlineRef
...
+
+struct HistoryInlineRef_LessThan
```
scunz: _If_ this is necessary at all, can it be done in private? (i.e. Move it to the .cpp)

History/HistoryEntry.h
```diff
@@ -61,9 +61,86 @@ struct HistoryInlineRef
     bool        mIsRemote   : 1;
     bool        mIsTag      : 1;
     bool        mIsStash    : 1;
+    bool        mIsDetached : 1;
+
+    bool operator ==(const HistoryInlineRef& other) const
```
scunz, line 66: What is the purpose of these operators? It looks like they're actually just implementing default behaviour.

History/HistoryEntry.h
```diff
@@ -61,9 +61,86 @@ struct HistoryInlineRef
...
+    {
+        if ( this == &other ) {
+            return true;
+        }
+
+        return  mRefName    == other.mRefName       &&
+                mRefName    == other.mRefName       &&
```
scunz: Despite the fact that it compares `mRefName` twice. :)

History/HistoryDiff.cpp
```diff
@@ -52,6 +52,8 @@ HistoryDiff::HistoryDiff()
     mToolbar = new QToolBar;
 
     mDiffTo = new QComboBox;
+    mDiffTo->addItem( trUtf8( "All parents" ), DTT_AllParents );
```
scunz, line 55: This looks like a good workaround for the missing "make this setting persistent somewhere" bug :nail_care:

History/HistoryEntry.h
antis81: Fixed :)

History/HistoryModel.cpp
```diff
@@ -350,28 +356,52 @@ void HistoryModel::scanInlineReferences()
             inlRef.mIsCurrent = false;
             inlRef.mIsStash = false;
         }
-        else if (ref == QLatin1String("refs/stash")) {
-            inlRef.mRefName = trUtf8( "<recent stash>" );
+        else if (ref == QLatin1Literal("refs/stash")) {
+            inlRef.mRefName = tr( "<recent stash>" );
             inlRef.mIsBranch = false;
-            inlRef.mIsCurrent = true;
+            inlRef.mIsCurrent = false;
```
scunz: That's wrong: _this_ stash is the most recent ones (You'll find the others be following the ref log).

History/HistoryEntry.h
antis81, line 66: Answer: Try removing this and write something like `if ( refA == refB ) ...`. The compiler fails in this case. Or am I missing something?

History/HistoryModel.cpp
```diff
@@ -350,28 +356,52 @@ void HistoryModel::scanInlineReferences()
...
+            inlRef.mIsStash = false;
+        }
         else {
             // qDebug() << "HistoryModel::scanInlineReferences => Unhandled ref:" << ref;
             continue;
         }
 
-        if (!refsById.contains(refs[ref])) {
-            refsById.insert(refs[ref], HistoryInlineRefs());
+        const Git::ObjectId &oid = refs[ref];
```
scunz: `refs` is a `Git::ResolvedRefs`, a `Git::ResolvedRefs` is a `QHash`.

Now, `refs` can either be `const` (which is not the case here), then `QHash`'s `const` operator `[]` never returns a reference but always a copy!
Or it can be non-const (which is the case here), then `operator[]` won't return a `const` reference...

What will the compiler do with that? I'd assume that it _might_ chose to create a temporary object on the stack and point the `oid` reference to it. It might further chose that the temporary object isn't required anymore after this line...

History/HistoryEntry.h
antis81: Actually this is a result of "lots of changes" :) - Sane Reason: Because `std::sort()` runs over it. Is QVector good for sorting as well?

History/HistoryEntry.h
antis81: Yes it can. Why if? The alternative would be to write an `operator <()`.

History/HistoryModel.cpp
antis81: Yes sorry. Should have been aware of that!

History/HistoryEntry.h
scunz: Yes, an `operator<` would be more natural. But see may pending comment for another thing :)

History/HistoryModel.cpp
antis81: Ah ... that actually confused me because of its ambiguous meaning. I'll change it back. It shouldn't have any effect though...

History/HistoryEntry.h
antis81: Seems no commit is pending - I got ~10 :) - what do you mean?

History/HistoryModel.cpp
```diff
@@ -380,53 +410,26 @@ void HistoryModel::scanInlineReferences()
         HistoryInlineRefs newRefs = refsById.value( e->id() );
         HistoryInlineRefs oldRefs = e->refs();
 
-        if( !newRefs.count() )
+        // sort newRefs: no need to sort oldRefs as it is already sorted)
+        std::sort( newRefs.begin(), newRefs.end(), lt );
```
scunz: This sort thing is mega inefficient. Since std::sort cannot know anything about the used Qt container (which is the design, so it can be used by whatever container is compatible to the stl), it can only sort things in place.

However, there's two options:
1) We change the QList back to a QVector, then we'll basically have a C-style array of `HistoryInlineRef`, which is pretty memory effective (esp. since we don't modify that container once it is created). But sorting inside it means that the assignment operator of `HistoryInlineRef` has to be called 3 times for every swap. This can cause a lot of memory fragmentation. I'd rather avoid that for something that might be processed so frequently like this code.

2) We accept the change to QList. In that case, we're having one more indirection (QString + 64 bit most certainly doesn't fit into 64 bit). So, a `QList<HistoryInlineRef>` will essentially be a `QVector<HistoryInlineRef*>`. Adding to the first option, 2 or 3 `new` and `delete` calls for HistoryInlineRef, which has probably a size of 48 or 64 byte. Causing yet more memory fragmentation.

I'd assume that I was much more efficient to just sort the initial vector while creating it.

History/HistoryModel.h
```diff
@@ -104,6 +114,15 @@ private slots:
     void beforeAppend();
     void afterAppend();
 
+    /**
+     * @internal
+     * @see RM::EventInterface
+     */
+    ///@{
```
scunz: This will probably uglify the doxygen output a lot - unless used _really_ consistently (and doesn't do what is expected anyway).

RefsViews/Branches/BranchesModel.cpp
```diff
@@ -165,68 +171,110 @@ bool BranchesModel::hasChildren( const QModelIndex& parent ) const
...
+
+    if (notify) {
+        endInsertRows();
+    }
+}
+
+RefScope* BranchesModel::scopeForRef(const Git::Reference& ref) const
+{
+    RefItem* scope = NULL;
+    if ( ref.isLocal() )        scope = mRoot->children[0];
```
scunz: These magic indicies are not exactly very future proof. Better keep some local variables with their values in the class...

RefsViews/Branches/BranchesModel.hpp
```diff
@@ -50,6 +57,23 @@ class BranchesModel : public QAbstractItemModel
 signals:
     void gitError( const Git::Result& error );
 
+private slots:
+    /**
```
scunz: This documentation stuff should definitely go to the .cpp files.

RefsViews/Branches/BranchesModel.cpp
```diff
@@ -165,68 +171,110 @@ bool BranchesModel::hasChildren( const QModelIndex& parent ) const
...
         Git::Result r;
-        Git::ReferenceList sl = repo.allReferences( r );
+        Git::ReferenceList sl = gitRepo.allReferences( r );
         if( !sl.isEmpty() )
         {
-            RefScope* scopeLocal = new RefScope( mRoot, tr( "Local" ) );
-            RefScope* scopeRemote = new RefScope( mRoot, tr( "Remote" ) );
-            RefScope* scopeOther = new RefScope( mRoot, tr( "Tags" ) );
-
             for( int i = 0; i < sl.count(); ++i )
```
scunz, line 212: This whole loop _should_ be removed (As soon as possible).

antis81, line 212: Actually I expect the `refCreated...` events to be fired here. For some reason they are not. If so, you are right. This would `double` the list.

History/HistoryModel.cpp
antis81: Fixed!

History/HistoryEntry.h
scunz, line 66: Why should the compiler fail on that? It will (just like with member-wise copy and member-wise assignment) do a member-wise compare. It will only have problems for more complex operators like less-that, exclusive-or-with-self-assignment etc... :)

History/HistoryEntry.h
```diff
@@ -61,9 +61,82 @@ struct HistoryInlineRef
...
+        return mRefName < other.mRefName;
+    }
+
+    bool operator <(const HistoryInlineRef& other) const
+    {
+        // sort order:
+        // - tag
+        // - detached HEAD
+        // - current branch
+        // - local branch
```
scunz, line 101: Thinking about this a bit more, I'd like to insert "Remote Tracking Branch" into the sort order between current branch and local branches - Would that make sense?

antis81: Btw.: Isn't it better to make the operator static? Or is it static by definition?

antis81, line 101: > Thinking about this a bit more, I'd like to insert "Remote Tracking Branch" into the sort order between current branch and local branches - Would that make sense?

Yeah, that makes sense. Though it would look a bit weird. Actually this is worth another feature request to "get it right". See [here](https://redmine.macgitver.org/issues/238).

Btw.: Isn't it better to make the operator static? Or is it static by definition?

antis81:
@scunz That's the way aha, aha - I like it!! :smile:

Event detection for references is mostly complete. Basically nothing is missing. Just some possible optimizations:
- The "source repo" of an event is not checked due to the fact, we don't have that information available yet.
- If we check for the "HEAD" reference in `RM::Ref::selfRefresh`, we can add a `headMoved` event. This has the advantage, that the status-view will get less events.
- On quit, the event `refAboutToBeDeleted` is sent for every reference. Could be blocked here!
- Of course the 15 secs. refresh interval is annoying. But hey, it works!!!!! We need some mechanism though to trigger an "immediate update".

:boat: I'd like to merge soon. As always not everything is implemented.

Other bigger tasks worth a new PR:
- Adding commits to `HistoryView` on creation
- Migrate / Refresh the `WorkingTreeView`
- Refresh the `IndexView`

RefsViews/RefItem.hpp
```diff
@@ -83,10 +91,13 @@ class RefNameSpace : public RefScope
 class RefBranch : public RefItem
 {
 public:
-    RefBranch( RefItem* p, const QString& t, const Git::Reference &ref );
+    explicit RefBranch(RefItem* p, const Git::Reference &ref);
```
scunz, line 94: Why is this `explicit`?

RefsViews/Branches/BranchesModel.hpp
```diff
@@ -50,9 +56,25 @@ class BranchesModel : public QAbstractItemModel
...
+private slots:
+    void onRefCreated(RM::Repo* repo, RM::Ref* ref);
+    void onRefDestroyed(RM::Repo* repo, RM::Ref* ref);
+    void onRefMoved(RM::Repo* repo, RM::Ref* ref);
+
+private:
+    QModelIndex index(RefItem* item) const;
+
+    void insertRef(bool notify, const Git::Reference& ref);
+    inline RefItem* insertNamespace(const bool notify, RefItem* parent, const QString& name);
```
scunz: If you declare something `inline`, then actually _make_ it inline by putting it into the header. These methods simply are not inline.

RefsViews/Branches/BranchesModel.cpp
scunz, line 212: As a matter of fact, we'll have to deal with the situation that the creation of whatever type of model is populated does _not_ align with opening / reading the repository. I.e. the underlying ViewContext might have expired or the view is simply opened later. Or even just the _current_ state of the code.

The consequence of that is: We need to be able to feed (preferably a single view) with all events that lead to the current state. This makes this loop unnecessary. That's actually the reason, why this whole event thing isn't base solely on Qt signals/slots.

In fact, when thinking about this, it might turn out that a Qt signal / slot wrapper was actually a total-bullshit idea. Hooking into different levels of the `RepoManager` and requesting a full-update on creating that hook would probably be much saner.

RefsViews/Branches/BranchesModel.hpp
antis81: Thanks a lot for the hint! These lines are subject to change anyways as I really need to  do the migration now.

RefsViews/Branches/BranchesModel.cpp
antis81, line 212: Actually my comment was total bullshit :). The `refCreated` events are not fired on opening/reading the repo, because `repoEventsBlocked` prevents that for good reason.

Well ... what do you mean by "hook into levels of RepoManager". Isn't this the same thing in green?

scunz, line 212: Yes, opening by itself shouldn't fire any events, actually. Though I'm not sure if that decision was right. However....

What we need is a way to force the actual model ("RepoMan") to send us a bunch of events that allow us to populate a given view from scratch. Once we've made that View-Concept of libBlueSky work, _it_ may decide to throw away all of our model (That we use to display) and asks us to recreate it when it is required.

What I was saying in my previous post, is that the Qt Signal/Slot mechanism doesn't allow us to do that. At least not for individual views. Let's say that we have 10 repositories open. Repository _A_ is currently displayed. Now, libBlueSky decides that repository _B_ has reached the maximum time allowed to be cached and just throws it away. The user now switches to either repository _B_ or to repository _C_ (which has never been open before). In both cases, libBlueSky will restore a ViewContext that doesn't have valid data assigned to it.

Now, have a look at the RepoMan, which is a huge tree of all open repositories and their individual objects (like References, Tags etc). What we can do with the Signal/Slot system is to _reset_ everything and have RepoMan send out events for everything. If every node in the RepoMan tree had the ability to have one or more observer(s) assigned to it (While the parent is implicitly an observer and event-proxy), we could hook up a ViewContext to any node inside the RepoMan hierarchy.

antis81, line 212: Basically I agree. I'm just practicing the "Art of small steps" here :) ... and I'm close to the finishing line - so I won't turn around at this point.

IOW:
- Step 1: Merge this PR, even if incomplete.
- Step 2: Currently the "RefModel" has nothing to do with RepoMan at all -> let's make them friends and Migrate "RefItem" to be actually aware of `RM::Ref` objects! This might become the observer you're talking about. I have local changes lying around here These became almost complete today morning. I'll open a new "PR" based on this merged state. Why? Because I don't want to destroy everything what's available here.
- Step 3: Make "RepoMan" the "Master-Model". This might be done in step 2, but seriously: let's discuss that [here](https://redmine.macgitver.org/issues/213). 

Btw.: I think the following "live-scenario" shows what you're describing about the "ViewContext". 

Just open a repository in MGV and close the "Repo-View" via the yellow `X` button. Then wait a few seconds and see what happens: The "Refs-View" will be suddenly cleared while the "HistoryView" stays populated. Didn't play around further, but it's saying enough ...

antis81:
:+1: Migration of "RefItem" works 90% locally! Some topics to be solved, but nothing too bad ... you'll see.

The new feature branch `ngf/migrate-refitem` is based on this one - which is unusual, but a sane solution for this situation.

This PR must be merged together with PR 22 in `libMacGitverCore` -> 
[look here](https://github.com/macgitver/libMacGitverCore/pull/22)

@scunz I'd like to merge this one. Could you do some last review?

@BugSniffer FYI

antis81:
@scunz Think we can close this one in favour to #99.

PR #98, opened 2015-02-09
migrate refitem to use RM::Ref

antis81:
The member `RefBranch->mRef` was a `Git::Reference` before. This is replaced now completely with a simple pointer to the valid `RM::Ref`. To avoid confusion I renamed the member to `mRefInfo`.

Known Bugs, that should be fixed, before the this PR is merged:
- [ ] Remote branches are missing in the tree
- [x] Scoped Branches are not shown.
- [ ] HEAD refs are not marked "orange" in submodules.
- [x] the prefixes "refs/heads", "refs/tags", ... should be hidden
- [ ] Regression: By accident, the reference name in the `Rename-Dialog` is initialized with the "pure" reference name, which is is wrong!

Here's a visual example, of how that currently looks:
mgv-migrate-refitem

### DIFF
--- a/RefsViews/Branches/BranchesModel.cpp
+++ b/RefsViews/Branches/BranchesModel.cpp
@@ -246,6 +246,9 @@ void BranchesModel::findInvalidRefItems(QVector<RefItem*>& invalidItems, RefItem
 ///@{
 void BranchesModel::onRefCreated(RM::Repo* repo, RM::Ref* ref)
 {
+    if ( repo != mData->repository() ) {
+        return;
+    }
     Git::Result r;
     Git::Reference gref = repo->gitRepo().reference( r, ref->fullName() );
     Q_ASSERT( r );
@@ -255,6 +258,10 @@ void BranchesModel::onRefCreated(RM::Repo* repo, RM::Ref* ref)
 
 void BranchesModel::onRefDestroyed(RM::Repo* repo, RM::Ref* ref)
 {
+    if ( repo != mData->repository() ) {
+        return;
+    }
+
     // TODO: This is an ugly workaround to find a matching RefItem!
     // We simply recursively search for invalid objects and delete them.
     QVector<RefItem*> invalidItems;
@@ -272,8 +279,11 @@ void BranchesModel::onRefDestroyed(RM::Repo* repo, RM::Ref* ref)
 
 void BranchesModel::onRefMoved(RM::Repo* repo, RM::Ref* ref)
 {
-    Q_UNUSED( repo )
     Q_UNUSED( ref )
+
+    if ( repo != mData->repository() ) {
+        return;
+    }
 
     // TODO: scan for changes in RefItems instead of performing a full update.
     QVector<int> updateRoles;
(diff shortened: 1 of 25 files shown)


RefsViews/Branches/BranchesModel.cpp
```diff
@@ -199,21 +196,14 @@ void BranchesModel::rereadBranches()
...
-        Git::Result r;
-        Git::ReferenceList sl = gitRepo.allReferences( r );
-        if( !sl.isEmpty() )
-        {
-            for( int i = 0; i < sl.count(); ++i )
-            {
-                const Git::Reference &currentRef = sl[ i ];
-                insertRef( false, currentRef );
-            }
+        foreach ( RM::Branch* info, repo->branches()->childObjects<RM::Branch>() ) {
```
scunz: Well, I can't see any code where you're actually asking RepoMan for the "Scope" objects :)

antis81: See today's implementation :). Thanks a lot!!!

scunz:
Did I make wrong naming decisions when I wrote the reference name analyser or did you confuse the concept of namespaces with scopes? Or maybe both? Will check that :)

antis81:
I remember we had once a really big discussion about this ... :). In general the names "namespace" and "scope" are historical. At the time I wrote the `RefItem` code, I didn't know about the namespacing mechanism in Git - the world was simple :).

In the MGV-Module `RefsViews` we currently have two classes relevant to this topic:
1. `RefScope`
2. `RefNamespace`

As `Git::RefName` uses similar names, these names became ambiguous and cannot be translated 1:1.

Another problem:
Both classes are technically 100% identical. They just serve different appearance purposes (display roles). I'll leave this fact aside and go straight forward to get the names clear once and for all.

Let's rename the classes to their meaning:
- RefScope             -> `HeaderItem`
- RefNamespace    -> `FolderItem`

Now it is more clear:
Both classes have nothing in common with a "scope" or "namespace" from `Git::RefName`! Actually a FolderItem can contain one of both.

Let's further simplify it:
The `HeaderItem` has a static name and groups references of the same type. So, for this topic, we don't need that and only talk about the `FolderItem`, which is built out of the reference `shorthand path`.

Here I had two problems:
When I understand `Git::RefName::scopes()` right, this list should be the separated shorthand path without the reference name. Seems like this is not implemented though and `Git::RefName::scopes()` is always empty. @scunz Could you check that, please?

Now why did I do the `Git::RefName::namespaces()` thing? Simple, because `Git::RefName::shorthand()` gives me an empty list, if `Git::RefName::isNamespaced() == true`.

Hope this "explanation" helps. Actually I plan to at least rename the `RefItem` classes here to remove the ubiquitous confusion with those names.

scunz:
Let me have my try with this code. I've actually refactored a lot already.

antis81:
I'll close this PR, as the implementation in #99 has quite some advantages compared to this. The idea is pretty much the same and I don't want to produce just work :).

